PR #15, opened 2019-08-19
Solution to Problem-1 and Problem-2 - Please Review

Arpan61:


Problem2.py
```diff
@@ -0,0 +1,33 @@
+# Time Complexity : O(n)
+# Space Complexity :O(n)
+# Did this code successfully run on Leetcode :Yes
+# Any problem you faced while coding this : Yes, the program is not optimal.
```
kaurramanpreet: you can optimize this using hashmaps

PR #16, opened 2019-08-19
Completed Tree1 : Please review

hrshah5:


Problem2.java
```diff
@@ -0,0 +1,36 @@
+// Time Complexity : Need help
```
kaurramanpreet: O(n) is the time complexity for your solution
as you are iterating over the preorder array once!

Problem1.java
```diff
@@ -0,0 +1,34 @@
+// Time Complexity : Need help
```
kaurramanpreet: O(n) where n is the number of elements is the time complexity for your code as you are iterating over all the elements of the tree!!

PR #22, opened 2019-09-09
Is Valid BST using recursion. Verified on LC:

manishjangid:


manishjangid:
@kaurramanpreet  : Do you have any suggestion here to make the indexes simpler. The reason is : in other high level languages for eg : in Java, there are functions where we can specify first and last index of the source array and it just copies it after allocating a new array.
Here while I was implementing it, it becomes challenging to get all the indexes right in the first attempt.
Thanks

kaurramanpreet:
Manish
Even when we use those functions in high level languages this work is done internally
So if there is no such function in c you have to access the elements this way only 
Or you can do one more thing
Instead of making 4 new arrays you can pass these indices as parameter(make separate function in that case) and then do the required process using those indices

kaurramanpreet:
Both your codes are absolutely  correct 
Well done

manishjangid:
Sure. I guess I will write a helper function then so that I can reuse it later in other programs.

PR #23, opened 2019-09-09
Completed-"Kindly review the code"

DeekanshaTandon:


validateBST.java
```diff
@@ -0,0 +1,49 @@
...
+
+        /* otherwise check the subtrees recursively
+        tightening the min/max constraints */
+        // Allow only distinct values
+        return (isBST(node.left, min, node.val) &&
+                isBST(node.right, node.val, max));
+    }
+}
+
+//Not sure why this solution doesn't work here
```
kaurramanpreet: because when we say valid bst that not only means that we need to have root.left<root.val and root.right>root.val
that also says that all the values on its left subtree must be smaller than that and on right subtree must be greater than that
and here you are returning at the moment when you get either of the conditions truee which is wrong

PR #59, opened 2020-01-01
submitting Trees-1

samhithaaaa:


buildTreeinorderpreorder.py
```diff
@@ -0,0 +1,22 @@
+class Solution(object):
+    def buildTree(self, preorder, inorder):
+        """
```
neha79091: Please write complexity

PR #78, opened 2020-02-24
Initial update for validation of BST from a BT

saiswaroopreddy:


Trees_1.1.java
```diff
@@ -0,0 +1,32 @@
...
+class Solution {
+    public boolean isValidBST(TreeNode root) {
+        return recursive(root, null, null);
+    }
+    
+    public boolean recursive(TreeNode node, int low, int up){
+        if(node == null)return true;//ending condition
+        
+        if(low ! = null && node.val <= low)return false;
+        if(up ! = null && node.val >=up )return false;
```
Lohith-Aswathappa: low and up are declared as an int in recursive method and they cannot be compared to null 
check again!

Lohith-Aswathappa:
check again and Fix the error

PR #84, opened 2020-02-25
Trees-1

Pavanisoma:
Throwing error in validate binary search tree (BST) code. Not sure what went wrong. Also please review time and space complexities

### DIFF
--- a/Validate Binary Search Tree.java
+++ b/Validate Binary Search Tree.java
@@ -1,0 +1,33 @@
+// Time Complexity :o(log n)
+// Space Complexity :o(n)
+// Did this code successfully run on Leetcode : No
+// Any problem you faced while coding this :
+//Yes, As per my understanding this code should work but it is failing testcases
+
+// Your code here along with comments explaining your approach
+/**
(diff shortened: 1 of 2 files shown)


Validate Binary Search Tree.java
```diff
@@ -0,0 +1,33 @@
...
+        int min= Integer.MIN_VALUE;
+        int max= Integer.MAX_VALUE;
+        return Validsearch(root,min,max);
+    }
+    public boolean Validsearch(TreeNode node,int minVal,int maxVal){
+        if(node ==null){return true;}
+        if(node.val >= minVal || node.val<= maxVal){
+            return  Validsearch(node.left,minVal,node.val) && Validsearch(node.right,node.val,maxVal); 
+ 
+        }
```
Lohith-Aswathappa: change the int to long data type & also in isValidBST
compare only less than or greater than, remove equal
and in second if, the logical operator should be && not ||

make these changes, it should be working. Let me know if there's any I missed out

Lohith-Aswathappa:
Change the code accordingly and submit the updated code

PR #102, opened 2020-04-07
Problem-1 completed

schakrad:


Problem-1.txt
```diff
@@ -0,0 +1,42 @@
+#Problem-1
+#TC:O(2^N)
```
devashishnyati, line 2: Why 2^n? You are visiting each node only once.

devashishnyati:
Good job Sindhu!
Please verify the time complexity.
Also submit the other problem.

PR #109, opened 2020-04-07
Tree 1

vaidyapawan:
106. Construct Binary Tree from Inorder and Postorder Traversal 
98. Validate Binary Search Tree

### DIFF
--- a/98. Validate Binary Search Tree
+++ b/98. Validate Binary Search Tree
@@ -1,0 +1,33 @@
+Time complexity: O(h)
+Space complexity: O(1)
+Did this code submitted succesfully on leetcode: Yes
+
+/**
+ * Definition for a binary tree node.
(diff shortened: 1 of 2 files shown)


98. Validate Binary Search Tree
```diff
@@ -0,0 +1,33 @@
+Time complexity: O(h)
```
devashishnyati, line 1: It should bt O(n)

98. Validate Binary Search Tree
```diff
@@ -0,0 +1,33 @@
...
+Space complexity: O(1)
```
devashishnyati, line 2: It should be O(h) -> Recursion stack

devashishnyati:
Good job Pawan!
Please verify the time and space complexities!

PR #111, opened 2020-04-08
Completed Trees 1 Please Review

kannan-mrugapphan:
passes all test cases on leetcode


BinaryTreeFromInorderPreorder.java
```diff
@@ -0,0 +1,36 @@
+//105.
+//time - O(n) - every element in prorder or inorder is visisted only once
```
devashishnyati, line 2: Searching for root index for n roots. Shouldn't this be O(n^2)?

devashishnyati:
Good job Kannan!

PR #116, opened 2020-05-11
Solved problems, long explanation sorry

domzhaomathematics:


domzhaomathematics:
Actually, O(n) space complexity instead of O(n), since not balanced

sowmyadvn:
> Actually, O(n) space complexity instead of O(logn), since not balanced

@domzhaomathematics For which problem is this? Can you please update space complexity in the PR instead?

sowmyadvn:
Both the solutions are correct, but please update space complexities of both the solutions

sowmyadvn:
Both the solutions are correct, but please update space complexities of both the solutions

PR #123, opened 2020-05-13
 Completed Trees -1 Please Review

jeelchristian:


40#Validate.txt
```diff
@@ -0,0 +1,24 @@
+// Time Complexity :O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this :No
+
+class Solution {
+    public boolean isValidBST(TreeNode root) {
+        return dfs(root,null,null);
+    }
```
super30admin, line 9: Code is not clean here in whole dfs approach

super30admin:
Great job with second solution. You have used pointer on both arrays and eliminated the need to hashmap

PR #126, opened 2020-05-14
Add files via upload

jaynirmal15:


ConstructBinaryTreefromInorderandPreorderTraversal.java
```diff
@@ -0,0 +1,47 @@
+package s30Coding;
+
+import java.util.*;
+
+//Time Complexity :- O(n) where n is number of elements in the list
+//Space Complexity :- O(h) where h is height of the tree
```
sowmyadvn, line 6: Map is auxiliary space here, isn't it? And we are storing all the nodes with indices in map. So, O(N) space complexity

sowmyadvn:
Awesome work Jay! Please check inline comment.  
Try to implement iterative approaches for both problems too.

sowmyadvn:
Awesome work Jay! Please check inline comment.  
Try to implement iterative approaches for both problems too.

PR #142, opened 2020-06-15
Completed Trees-1: Please review

shantanu609:


AbhiniveshP:
Nice work Shantanu !!!
In the question isValidBST() for iterative approach where you used Stack, the Space would be O(h) which varies with the height of the BST and not O(n). 
In the question buildTree(), on line 16 where you used .index() to find the index that would separately take O(n) time which makes your algorithm to be O(n^2). So, try using a dictionary there.

AbhiniveshP:
Nice work Shantanu !!!
In the question isValidBST() for iterative approach where you used Stack, the Space would be O(h) which varies with the height of the BST and not O(n). 
In the question buildTree(), on line 16 where you used .index() to find the index that would separately take O(n) time which makes your algorithm to be O(n^2). So, try using a dictionary there.

shantanu609:
Please check. I have updated the solutions and comments based on your inputs.

PR #169, opened 2020-06-18
Finished. Please review.

srichaitrakareddy:


iivaishnavii:
Good job, Chaitra. You got the solution,TC and SC right! Nice to see you try out multiple approaches

PR #175, opened 2020-06-20
Create ValidateBinaryTree.py

Shashank001122:


Lohith-Aswathappa:
Nice try, Shashank!
Try the other problems as well.

PR #180, opened 2020-06-24
Added Problem 1 and 2

auluckh23:


AbhiniveshP:
For problem 1, space complexity would be O(h). Try implementing iterative approach for problem 2 as well.

AbhiniveshP:
For problem 1, space complexity would be O(h). Try implementing iterative approach for problem 2 as well.

PR #198, opened 2020-07-30
Solutions for Trees-1 module

Basavsagar:


problem-1.py
```diff
@@ -0,0 +1,47 @@
...
+
+1. Kep 2 global variables isValid and prev. prev contains the element which was last visited in the in-order traversal
+2. we know that in order traversal of BST will be in ascending order, so we keep chekcing the current element with the prev element
+3. if pre element is >= current element then we make the global variable isValid as False and return
+4. finally we return the gloabl variable isValid
+
+"""
+
+# Time Complexity : O(n) 
+# Space Complexity : O(1)
```
Sumesh-H, line 15: O(h) height of the tree.

Sumesh-H:
Good work, Basavsagar. In Tree problems, we consider the height of the tree for Space complexity in the final analysis.

Basavsagar:
Okay got it.. so if we are using recursion then we say space complexity is O(h) where h is height of the tree and its log(n) ?

Sumesh-H:
> Okay got it.. so if we are using recursion then we say space complexity is O(h) where h is height of the tree and its log(n) ?

Yes. Under the hood, its O(log n), at final analysis just specify O(h).

PR #254, opened 2020-10-03

DhruvJawalkar:


TreeFromInorderAndPreorder.java
```diff
@@ -0,0 +1,46 @@
+// Time Complexity : O(n), we would process and build one root node at a time I guess
```
DhruvJawalkar, line 1: I wasn't very clear about how to calculate the TC using the master theorem here

TreeFromInorderAndPreorder.java
```diff
@@ -0,0 +1,47 @@
+// I wasn't very clear about how to calculate the TC using the master theorem here
+// Time Complexity : O(n), we would process and build one root node at a time 
+// Space Complexity : O(n) + O(log n) for (preorder, inorder, hashmap) + for call stack 
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this :
+// This was hard, I did not get the intuition to solve the problem, I was trying to make the tree by traversing the array in reverse given right would be the rightmost element
```
kannan-mrugapphan, line 6: The other approach where you have a separate list at each call will build more intuition. I would recommend solving that as well.

kannan-mrugapphan (commented):
Nice Work. good Explanation

kannan-mrugapphan:
Nice Work. good Explanation

PR #259, opened 2020-10-06
Trees-1 only one completed

srinidhib-cmd:
Hi, I was able to come up with logic for only one of the questions. 
I will submit the second question after class.


Lohith-Aswathappa (commented):
Good work, Srinidhi!
Try the other problem as well.

Lohith-Aswathappa:
Good work, Srinidhi!
Try the other problem as well.

PR #267, opened 2020-10-06
Trees-1

Hanishsai:
Done with Problem 1 and Problem 2


Problem1.java
```diff
@@ -0,0 +1,35 @@
+// Time Complexity :O(N)
+// Space Complexity : O(1)
```
kannan-mrugapphan, line 2: Space become O(h) due to call stack

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #287, opened 2020-11-03
validate if BST

harshasanjeeva:


Lohith-Aswathappa (commented):
Good work, Harsha!
Try the other problem as well.

Lohith-Aswathappa:
Good work, Harsha!
Try the other problem as well.

PR #294, opened 2020-11-05
Submitting 2 solutio

sumitjanawlekar:
Submiting 2 solutions each for both the problems
 construct_binary_tree_from_preorder_and_inorder --> brute force and optimal solution
validate_binary_tree --> iterative and recursive solution


validate_binary_search_tree recursive.java
```diff
@@ -0,0 +1,56 @@
+// Time Complexity : O(n), where n is the number of nodes in the tree
+// Space Complexity :O(n), where n is the height of the tree, to maintain stack
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : not clear, if how function level return terminates the program, but base case level return
```
kannan-mrugapphan, line 4: Its like 3(parent) calling 2 calling 1 (base). Return on base case terminates 1 thus control returns to 2. Now return in the recursive case terminates 2 and control returns to 3. The recursive case return terminates 3 and control transfers to driver or main. If still unclear, please DM

validate_binary_search_tree recursive.java
```diff
@@ -0,0 +1,56 @@
+// Time Complexity : O(n), where n is the number of nodes in the tree
+// Space Complexity :O(n), where n is the height of the tree, to maintain stack
```
kannan-mrugapphan, line 2: Space is O(h) where h is tree's height

construct_binary_tree_from_preorder_and_inorder_traversal.java
```diff
@@ -0,0 +1,66 @@
+// Time Complexity : O(n), where n is the number of nodes in the tree (time requried to copy elements from inorder array to hashMap)
+// Space Complexity : O(n), O(n), where n is the number of nodes in the tree (space required for hashMap)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : I was trying to pass the preorderIndex as a local variable instead of global in recursion,
```
kannan-mrugapphan, line 4: preIndex is a primitive which is modified in every call. So it cannot be passed directly. If you do so, the variable won't retain expected value. If you want to pass it, you can create an array of size 1 and pass the array (pass by reference)

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #313, opened 2020-12-09

zaibafathima-9:


kannan-mrugapphan (commented):
Good Job. Good work with the brute force for both problems.

kannan-mrugapphan:
Good Job. Good work with the brute force for both problems.

PR #335, opened 2021-01-07
completed! - Please review

Helisha1:


ramya-sruthi-d (commented):
Good job, in the second problem you are validating the tree so you are going through everynode  at least once. So, the time complexity is O(num of nodes).

ramya-sruthi-d:
Good job, in the second problem you are validating the tree so you are going through everynode at least once. So, the time complexity is O(num of nodes).

PR #343, opened 2021-01-12
Completed. Please review.

Faraaz1994:


maithrreye:
Good Work Faraaz ! isn't the TC of the BuildTree is O(N) ?

maithrreye:
Good Work Faraaz ! isn't the TC of the BuildTree is O(N) ?

PR #401, opened 2021-02-10
Completed Sprint Trees1 - Please Review

JanhaviDahihande:


binary_tree_from_preorder_inorder.py
```diff
@@ -0,0 +1,34 @@
+'''
+Time complexity: O(2n*n)
+Space complexity: O(max_depth)
```
nsibal, line 3: Since you are creating sub_arrays as well, I believe your SC would be bigger than this.
You can do way better than O(n^2) TC. Please improve your solution.

nsibal (commented):
Validate BST looks good. Read inline comments for Build Tree from PreOrder and InOrder.

nsibal:
Validate BST looks good. Read inline comments for Build Tree from PreOrder and InOrder.

PR #402, opened 2021-02-11
Trees-1  Completed

singhnavraj24:


sample.java
```diff
@@ -0,0 +1,108 @@
...
+
+        //if we reach here then we have processed the whole tree and 
+        //it is BST
+        return true;
+    }
+}
+
+//Problem 1
+//Time: O(n)
+//Space: O(depth of Tree)
```
nsibal, line 50: SC would be O(n) because you're building a HashMap of size n.

nsibal (commented):
Looks good! Read inline comments.

nsibal:
Looks good! Read inline comments.

PR #416, opened 2021-03-09
Completed Problem1 : Please review!

mannatDev:


Problem-2.java
```diff
@@ -0,0 +1,45 @@
+Time complexity : O(n)
```
nsibal, line 1: Time complexity seems incorrect. You are searching for the index in the array. At the worst, this algorithm runs in O(n^2) time.

nsibal (commented):
Read inline comments about TC, otherwise looks good!

nsibal:
Read inline comments about TC, otherwise looks good!

PR #421, opened 2021-03-09
Completed problems. Please review.

dkeerthireddy:


BTusingPreorderInorder.java
```diff
@@ -0,0 +1,58 @@
+//Time Complexity: O(n)
```
nsibal, line 1: TC won't be O(n) because you are searching for the index in each recursive call. At the worst, this would run in polynomial time.

nsibal (commented):
Read inline comments, looks good otherwise!

nsibal:
Read inline comments, looks good otherwise!

PR #426, opened 2021-03-11
Completed Trees-1

aswathimanden:


Problem_2.java
```diff
@@ -0,0 +1,27 @@
+// Time Complexity : O(n), n= no. of nodes in tree
```
nsibal, line 1: Since you are running a `for` loop at each recursive call, this algorithm won't run in linear time.

nsibal (commented):
Read inline comments, otherwise looks good!

nsibal:
Read inline comments, otherwise looks good!

PR #482, opened 2021-06-16
added trees1

pandas5308:


maithrreye:
Please try valid BST with min-max or with inroder traversal checking with previous solution for optimal approach ! Finish the other problem too ! Good try

maithrreye:
Please try valid BST with min-max or with inroder traversal checking with previous solution for optimal approach ! Finish the other problem too ! Good try

PR #611, opened 2021-09-15

Varshith6:


Pinak08 (commented):
SC for valid binary search tree will be O(height).
As you are using DFS in it so there is an inbuilt stack being used for it.
Also for question 1 tc will not be n^2 but n.

Pinak08:
SC for valid binary search tree will be O(height).
As you are using DFS in it so there is an inbuilt stack being used for it.
Also for question 1 tc will not be n^2 but n.

PR #697, opened 2022-01-11
chore - provided optimal solution 

BitLegion1996:


sonalgupta276 (commented):
Good job! Do optimize solutions as well. Mention TC and SC too

sonalgupta276:
Good job! Do optimize solutions as well. Mention TC and SC too

PR #747, opened 2022-03-01
Trees 1 solved - pls review

tusharachigi:


sanjayvarma199 (commented):
Good Job!! Please also include TC and SC

sanjayvarma199:
Good Job!! Please also include TC and SC

PR #751, opened 2022-03-03
Completed Trees-1 _ Please Review

suhaas-livcd:
- Worked on two problems, please review

### DIFF
--- a/ValidateBST.java
+++ b/ValidateBST.java
@@ -1,5 +1,5 @@
-// Time Complexity : O(h) height of tree, if skewed then worst case iterate all
-// Space Complexity : Not using additional list to store, just constant variables
+// Time Complexity : O(n) ~ O(h) h - height of tree because if skewed then worst case visit all nodes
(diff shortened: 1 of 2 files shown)


ValidateBST.java
```diff
@@ -0,0 +1,68 @@
+// Time Complexity : O(h) height of tree, if skewed then worst case iterate all
+// Space Complexity : Not using additional list to store, just constant variables
```
shrivatsav: Please check the Time and Space complexity.

shrivatsav (commented):
Good Work! Please address the feedback.

shrivatsav:
Good Work! Please address the feedback.

ValidateBST.java
suhaas-livcd: @shrivatsav Thank you for the review.
Updated the complexity for ValidateBST.java.
// Time Complexity : O(n) ~ O(h) h - height of tree because if skewed then worst case visit all nodes
// Space Complexity : O(n) function stack size; depth of tree nodes. If no function stack size considered then O(1) for constant variables

PR #768, opened 2022-03-31
Completed Trees-1 - Please review.

het3008:


BST.java
```diff
@@ -0,0 +1,59 @@
+/*
+Time Complexity : O(n) where n is time taken to find whether it's BST or not.
```
shrivatsav, line 2: Please check what `n` indicates.

Inorder.java
```diff
@@ -0,0 +1,48 @@
+
+/*
+Time Complexity :  O(n^2) where finding the index is O(n) and buildig the arrays is O(n).
+Space Complexity: O(1) as we are not using any extra auxillary space.
```
shrivatsav: Please try to optimize on time. Please check the space complexity.

shrivatsav (commented):
Good Work! Please address the feedback.

shrivatsav:
Good Work! Please address the feedback.

Inorder.java
het3008: I am confused what is space complexity in this. Could you please guide me?
As per me it is O(n).

BST.java
het3008: Here n means, the amount of time we had taken to find whether it's BST or not. suppose If it's BST then we have to traverse all the elements in the tree.

het3008 (commented):
Reduced time complexity from O(n^2) to O(n).
Please review.

Inorder.java
shrivatsav: You are using  `Arrays.copyOfRange()` here. So this will need additional space.

het3008: Okay, got it.
Thank You.

PR #854, opened 2022-06-17
Assignment Complete.

vraj9845:


Construct Bin. Tree from inorder,preorder/Solution.java
```diff
@@ -0,0 +1,24 @@
+// Time Complexity: O(n)
```
ishika2741: Kindy check the time complexity. Every time finding the root will take  O(n^2) .

ishika2741 (commented):
Well done!

ishika2741:
Well done!

Construct Bin. Tree from inorder,preorder/Solution.java
vraj9845: Okay changes made

PR #935, opened 2022-08-05
Completed Trees-1-Please Review

rajdshah:


ValidBST.py
```diff
@@ -0,0 +1,34 @@
+# TC : O(N)
+# SC : O(N)
```
jeelpatel18, line 2: For DFS SC will be O(H). Height of the tree but, you are using list so that's why it is O(N)

jeelpatel18: Plus for loop is having O(N) so you can say like N+H

jeelpatel18: But if you check the conditions for value is lesser than its previous element while doing inorder(Tree Traversal) then you don't need to store the tree nodes in list

jeelpatel18 (commented):
Good job.

jeelpatel18:
Good job.

PR #1003, opened 2022-10-10
Validate BST

karan137:


iterate BST
```diff
@@ -0,0 +1,77 @@
...
+        bool case2 = minmaxcheck(root->right, root->val , max);
+        
+        return case1&&case2;
+        }
+    
+    bool isValidBST(TreeNode* root) {
+        if(root == NULL ){
+            return true;
+        }
+    return minmaxcheck(root, INT_MIN, INT_MAX);
```
rajchoksi1997, line 75: comment: 1 you are setting min value to INT_MIN and max value to INT_MAX hence when root-> val == INT_MIN || INT_MAX your code is failing because of the if condition 
refer comment 2

comment 3:
hence you should initialize min and max values as treenode pointers and keep a null check in your if condition
then your code will work

iterate BST
```diff
@@ -0,0 +1,77 @@
...
+ */
+class Solution {   
+    public:
+    bool minmaxcheck(TreeNode* root, int min, int max){
+        if(root == NULL){
+            return true;
+        }
+        
+        
+        if(root->val >= max || root->val <=min){
```
rajchoksi1997, line 60: comment 2: in this condition you are check if value is equal to max or min value so here your code fails.

rajchoksi1997 (changes requested):
make the requested changes and add space and time complexities

rajchoksi1997:
make the requested changes and add space and time complexities

PR #1014, opened 2022-10-21
Completed Sprint Trees-1 : Please Review

jaszaveri:


Question1.py
```diff
@@ -0,0 +1,16 @@
+#Time-Complexity: O(n)
+#Space-Complexity: O(1)
```
daredevilteja, line 2: SC is wrong

daredevilteja:
Good Work! Keep it up

Question2.py
```diff
@@ -0,0 +1,19 @@
+#Time-Complexity: O(n^2)
```
daredevilteja, line 1: Please recheck TC and SC

daredevilteja:
Good Work! Keep it up

PR #1035, opened 2022-11-20
Create Sample.java

RiddhiJaju123:


rajchoksi1997 (approved):
space complexity would be o(h) instead of o(n)

rajchoksi1997:
space complexity would be o(h) instead of o(n)

PR #1066, opened 2023-01-11
completed trees-1

rohitbassi:


problem1.py
```diff
@@ -0,0 +1,20 @@
+# Time Complexity: O(N^2)
```
daredevilteja, line 1: recheck TC

daredevilteja (changes requested):

daredevilteja:
Good Work! Keep it up.

PR #1085, opened 2023-02-16
Completing Trees-1 again for practice

RadhikaKulk:


ValidateBinarySearchTree.java
```diff
@@ -0,0 +1,27 @@
+// TC: O(n)
+// SC: O(n) (considering the recursive stack space)
```
osandeepreddy, line 2: Good Job. SC= O(h)

osandeepreddy:
Good Job. SC= O(h)

PR #1152, opened 2023-06-26
Completed Trees 1

sana-h-pathan:


ishika2741 (commented):
Good work! Please add Tc and Sc . Also submit other problems as well.

ishika2741:
Good work! Please add Tc and Sc . Also submit other problems as well.

PR #1281, opened 2024-01-15
Added solutions for Trees-1

srawlani22:


validate_bst.py
```diff
@@ -0,0 +1,46 @@
+""" Explanation: We first recursively get the root of the bst, in the inorder traversal manner. Then we simply check if those values
+    traversed are in the ascending order. 
+    Time Complexcity: O(n)
+    Space Complexcity: O(1)
```
ishika2741, line 4: under hood Stack space would be there which would be O(H) where H is the height of the tree

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1295, opened 2024-02-13
Trees 1 (Java)

anishSingh7010:


Nandha2Tech (commented):
Great Work! Try to use HashMap to get the position from inorder  in the problem inorder  to preorder constrcution. This will reduce TC from O(n^2) to O(n).

Nandha2Tech:
Great Work! Try to use HashMap to get the position from inorder  in the problem inorder  to preorder constrcution. This will reduce TC from O(n^2) to O(n).

anishSingh7010:
Thanks for the review! Will use a hashmap here.

PR #1316, opened 2024-03-21
Tree-1 done

vaishnavimore23:


ConstructBinaryTree.py
```diff
@@ -0,0 +1,26 @@
+# TC = O(nk)
```
ishika2741, line 1: Try to submit the optmized solution as well

ValidateBinaryTree.py
```diff
@@ -0,0 +1,32 @@
+# TC = O(n)
+# TC = O(1)
```
ishika2741, line 2: You need to consider recursive stack space as well

ValidateBinaryTree1.py
```diff
@@ -0,0 +1,28 @@
+# TC = O(n)
+# TC = O(1)
```
ishika2741, line 2: Here as well recursive stack space

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1337, opened 2024-05-30
done

dewann:


Nandha2Tech (commented):
Great Work! TC for buildTreeInorderPreorder problem in the logic you coded out would be O(n^2) and SC would also be O(n^2). Try to implement in O(n) TC too.

Nandha2Tech:
Great Work! TC for buildTreeInorderPreorder problem in the logic you coded out would be O(n^2) and SC would also be O(n^2). Try to implement in O(n) TC too.

dewann:
optimized for On TC and Sc as suggested

PR #1371, opened 2024-07-11
Done Trees-1

ssd9881:


Nandha2Tech (commented):
Great Work! TC for the logic that you provided for BST from Preorder and Inorder would be O(n^2). Kindly try to do it in O(n) manner. And also TC for validate BST would be O(n) and SC would be O(h).

Nandha2Tech:
Great Work! TC for the logic that you provided for BST from Preorder and Inorder would be O(n^2). Kindly try to do it in O(n) manner. And also TC for validate BST would be O(n) and SC would be O(h).

PR #1386, opened 2024-07-26
completed Trees-1

dhikki-study:
########98. Validate Binary Search Tree######################################
// Time Complexity : n
// Space Complexity : O(1)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
We took previous pointer as NULL and started traversing tree in INORDER and making last visited node as PREV and current node as CURR and comparing always prev to be less than current 

        

#######105. Construct Binary Tree from Preorder and Inorder Traversal######################################
// Time Complexity : n
// Space Complexity : O(n)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : I coded without using index  and instead passing new ignorer and preorder array each time but than performed white boarding and did it by using index so extra space was used

// Your code here along with comments explaining your approach in three sentences only
We use index to traverse in preorder, so each index give us next node the problem is to identify which are left node which was helped using inorder by find the location of root in ignorer array and all element on its left are eft side of tree

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,91 @@
+########98. Validate Binary Search Tree########################################################################################################################
+// Time Complexity : n
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : no problem
+
+// Your code here along with comments explaining your approach in three sentences only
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #1402, opened 2024-09-21
Trees-1 Complete

divyameharwade:


Nandha2Tech (commented):
Great Work! The logic done here in buildTree uses inorder[:index] and inorder[index + 1:]. Slicing of the array creates another O(n) time. So TC here becomes (n^2). Kindly try to optimize it to O(n).

Nandha2Tech:
Great Work! The logic done here in buildTree uses inorder[:index] and inorder[index + 1:]. Slicing of the array creates another O(n) time. So TC here becomes (n^2). Kindly try to optimize it to O(n).

PR #1468, opened 2025-01-23
Solved 2 probs

joshi112:


Nandha2Tech (commented):
Great Work! Space complexity would be O(h) recursive stack space (h is the height of the tree).

Nandha2Tech:
Great Work! Space complexity would be O(h) recursive stack space (h is the height of the tree).

PR #1473, opened 2025-01-25
attempt at Trees-1

abhinavg-neu:


Nandha2Tech (commented):
Great Work! Kindly try to optimize construct binary tree from preorder and inorder to O(n) time complexity. (Try to store inorder array in hashmap so that search can be optimized to O(1))

Nandha2Tech:
Great Work! Kindly try to optimize construct binary tree from preorder and inorder to O(n) time complexity. (Try to store inorder array in hashmap so that search can be optimized to O(1))

PR #1504, opened 2025-02-12
finished homework-Trees-1

raaulc:


Erlybird:
Good job! Please mention the TC and SC of the solutions

Erlybird:
Good job! Please mention the TC and SC of the solutions

PR #1614, opened 2025-06-13
Completed Trees-1

RishithaReddyS:


super30admin:
Strengths:
- The use of a helper function with bounds is a good approach.

Areas for Improvement:
- The space complexity analysis could be more precise. It should be O(h) where h is the height of the tree, not O(n) in all cases.
- Avoid using `new int` for passing bounds to prevent memory leaks and improve performance. Consider passing values directly or using references.
- The comment about time and space complexity in the code could be updated to reflect the more accurate O(h) space complexity.

PR #1621, opened 2025-06-25
Trees-1 completed

itizsanket:


super30admin:
Strengths:
- Clear and concise implementation of both problems.
- Correct use of data structures (stack for iterative traversal, hashmap for index lookup).
- Good handling of edge cases (e.g., null root in Problem1, left > right in Problem2).

Areas for Improvement:
- For Problem1, the space complexity could be more precisely stated as O(h) where h is the height of the tree, as the stack size depends on the tree's height.
- In Problem2, the TreeNode class is imported but not defined in the solution. It would be better to include the TreeNode definition or ensure it's available in the context.
- Adding brief comments explaining the logic (e.g., why in-order traversal works for BST validation) would enhance readability for others.

PR #1625, opened 2025-07-06
Done Trees-1

Nav0708:


super30admin:
- Problem-1: Your solution is correct and efficient. The time complexity is actually O(n), not O(2n), as each node is visited exactly once. The space complexity is O(h) due to the recursion stack, not O(n). Your approach is clean and well-commented.
- Problem-2: Your solution correctly builds the tree but could be optimized by avoiding array copying. Instead, you could pass indices to the recursive function to mark the current subtree's boundaries in the preorder and inorder arrays. This would improve both time and space efficiency. The rest of the code is well-structured and easy to follow.

super30admin:
- Problem-1: The solution is correct and efficient. The explanation is clear, and the code is well-structured. Good job!
- Problem-2: The solution is correct but can be optimized. Instead of copying arrays, consider using indices to represent subarrays in the recursive calls. This would reduce both time and space complexity to O(n). Additionally, using a hashmap to store the indices of inorder elements can help reduce the time complexity of finding the root index from O(n) to O(1) per call.

PR #1643, opened 2025-07-27
completed

sejalrj:


super30admin:
- Problem 1.py: Excellent job! Your solution is correct, efficient, and well-structured. The use of a helper function with range checks is a clean and effective approach.
- Problem 2.py: Your approach is on the right track, but the solution is incomplete because the helper function is not called in the main function. Additionally, ensure that the main function returns the result of the helper function. The use of a hashmap for O(1) lookups is a good optimization. Consider adding comments to explain the logic, especially the use of a global index via a list.

PR #1652, opened 2025-08-17
Janhvi Guha Trees 1

Janhvi2000:


super30admin:
- Strengths:
  - Both solutions are correct and implement the required logic accurately.
  - Time and space complexities are correctly identified.
- Areas for Improvement:
  - For `construct_bt_pre_in.py`, consider using a hashmap to store inorder indices to reduce the time complexity from O(n^2) to O(n). This would be more efficient for larger trees.
  - While the current implementation works, it might be worth mentioning the trade-offs between the recursive approach and an iterative one (e.g., stack-based) for `validateBST.py`, especially in languages where recursion depth can be a concern.

PR #1654, opened 2025-08-18
Done Trees - 1

has1th:


super30admin:
- Problem1.java: Your solution is correct and efficient, but the space complexity could be improved. Consider mentioning the worst-case space complexity in your comments. Also, the iterative approach is good, but you might want to explore recursive solutions for comparison.
- Problem2.java: Excellent job! The solution is correct and efficient. The use of a hashmap for O(1) lookups is a great choice. Just ensure to document the space complexity more clearly in your comments.

PR #1660, opened 2025-08-25
Trees-1

botlamanoj29:
Executed on leet code both the problems.

### DIFF
--- a/preOrderTransversal.py
+++ b/preOrderTransversal.py
@@ -1,0 +1,25 @@
+# Time Complexity : It is O(n) since we are iterating with all nodes.
+# Space Complexity : It is O(h) considering the height of the tree in the stack.
+# Did this code successfully run on Leetcode :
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- The student correctly identified the time and space complexity for both problems.
- The recursive approach in both solutions is conceptually correct.

Areas for Improvement:
- In `validateBST.py`, the logic to validate the BST is missing. The student should compare the current node's value with the previous node's value during the inorder traversal.
- Avoid using class variables for state that should be local to the recursion. Instead, pass the previous value as a parameter or use a helper function with instance variables.
- In `preOrderTransversal.py`, the `preOrderTree` list is unnecessary and should be removed. The solution should focus solely on building the tree.
- Add comments to explain the logic and steps, especially in the recursive functions.
- Ensure that the code is tested thoroughly to catch logical errors.

PR #1682, opened 2025-10-11
Completed Trees-1

vvirani1908:


super30admin:
- Strengths:
  - Both solutions are correct and efficient.
  - The use of an iterative approach in problem1.py is a good alternative to recursion.
  - The hashmap in problem2.py for O(1) lookups is a smart optimization.
- Areas for Improvement:
  - In problem1.py, the space complexity could be noted as O(h) (height of the tree) in the best case, as the stack will only hold nodes along the path from the root to the leftmost leaf at any point.
  - In problem2.py, the `root.idx` assignment is unnecessary for the solution and could be removed unless it's for debugging purposes.
  - Consider adding more context about the problem in the comments for better clarity, especially for problem2.py, where the approach might not be immediately obvious to everyone.

PR #1685, opened 2025-10-25
Validate BST & Create Binary Tree from inorder and preorder

ashishbadhe97:


super30admin:
Strengths:
- The student has correctly implemented both problems with appropriate time and space complexity.
- The student has shown versatility by providing two different approaches for ValidateBST.

Areas for Improvement:
- For TreeFromInorder&Preorder, the student could consider adding a check to ensure the input arrays are not null or empty, though the problem constraints likely guarantee this.
- In ValidateBST, the boolean-based recursion method is more elegant and avoids the need for a class-level flag variable, which is a better practice. The student could consider using this approach primarily.
- The student could further optimize the TreeFromInorder&Preorder solution by avoiding passing the entire preorder and inorder arrays in each recursive call, though this is a minor point.

PR #1729, opened 2026-02-26
Complete Trees-1

PrakarshKamal:


super30admin:
Strengths:
- The solution correctly implements the recursive range-checking method for validating BST.
- It handles edge cases with integer boundaries by using long min and max.

Areas for Improvement:
- The TreeNode class should not be redefined in the same file. The problem already provides the definition, so including it again may cause duplication. Remove the TreeNode class from ValidateBST.java.
- Although the solution is correct, note that the space complexity is O(h) due to recursion, which is efficient. However, in the worst case (skewed tree), h could be n, leading to O(n) space. This is acceptable and matches the reference solution.
- The solution for a different problem (ConstructBTFromPreAndIn) is included in the same submission. Make sure to only submit the code for the problem being solved.

PR #1731, opened 2026-02-27
Solved Trees-1

pratul2789:


super30admin:
Strengths:
- You have implemented two correct approaches for validating a BST, showing good understanding of the problem.
- The active solution (range-based) is efficient and has optimal time and space complexity.
- The code is well-commented, making it easy to understand your logic.

Areas for Improvement:
- The file contains an unrelated solution (for building a tree from preorder and inorder). It's important to keep solutions for different problems separate to avoid confusion. For this problem, only the Validate BST solution should be present.
- In the range-based solution, the helper function parameters (minVal and maxVal) could be named more clearly to indicate they are the lower and upper bounds (e.g., low, high or min_bound, max_bound).
- Although not necessary, you could consider using an iterative approach for the in-order traversal to avoid recursion stack space, but the recursive solution is acceptable.

Overall, your solution for Validate BST is correct and efficient. Just ensure that you submit only the relevant code for the problem.

PR #1735, opened 2026-03-15
Add implementation for leetcode problems 98, 105

rishigoswamy:


super30admin:
### Validate BST (leetcode_105.py)
Strengths:
- You have implemented an efficient solution for constructing a binary tree from preorder and inorder traversals. The use of a hashmap for O(1) index lookup and the recursive approach with indices (avoiding list slicing) is optimal.
- The code is clean and well-commented, which is good practice.

Areas for Improvement:
- The main issue is that you solved the wrong problem. The problem you were assigned is "Validate BST", not "Construct Binary Tree from Preorder and Inorder Traversal". Please double-check the problem statement before starting to code.
- For the "Validate BST" problem, you need to implement a solution that checks if a given binary tree is a valid BST. The reference solution uses an in-order traversal to check if the sequence is strictly increasing. Alternatively, you can use a recursive approach with min and max bounds.
- In the future, ensure you are solving the correct problem by carefully reading the title and description.

VERDICT: NEEDS_IMPROVEMENT

---

### Construct Binary Tree from Preorder and Inorder Traversal (leetcode_98.py)
It appears that you have submitted a solution for the wrong problem (LeetCode 98) instead of the required problem (LeetCode 105). Please ensure that you are solving the correct problem. For LeetCode 105, you need to construct a binary tree from preorder and inorder traversal arrays.

For the correct problem, the key idea is to use the preorder array to determine the root node (the first element), then find the position of that root in the inorder array. This splits the inorder array into left and right subtrees. Recursively build the left subtree from the left part and the right subtree from the right part.

You can use a hash map to store the indices of the inorder array for efficient root position lookup. The reference solution provided in C++ can be a good guide. You'll need to adapt it to Python.

VERDICT: NEEDS_IMPROVEMENT

PR #1737, opened 2026-03-25
Trees-1

nagasai67:


super30admin:
### Validate BST (validateBST.py)
It seems there might have been a mix-up with the problem. The problem you solved is "Construct Binary Tree from Preorder and Inorder Traversal", but the problem you were asked to solve is "Validate BST". Please double-check the problem statement and ensure you are working on the correct one.

For the "Validate BST" problem, you need to check if a binary tree meets the BST properties: 
- The left subtree of a node contains only nodes with keys less than the node's key.
- The right subtree of a node contains only nodes with keys greater than the node's key.
- Both the left and right subtrees must also be binary search trees.

A common approach is to perform an inorder traversal and check if the sequence is strictly increasing. Alternatively, you can use recursion with min and max bounds to validate each node.

Here's a hint for the correct solution:
You can use a recursive helper function that takes a node and the current min and max allowed values. For the root, min can be -infinity and max can be +infinity. Then for the left subtree, update the max to be the current node's value, and for the right subtree, update the min to be the current node's value.

Please revisit the problem and try to implement the correct solution. If you have any questions, feel free to ask.

VERDICT: NEEDS_IMPROVEMENT

---

### Construct Binary Tree from Preorder and Inorder Traversal (constructTree.py)
Your solution is well-implemented and efficient. You correctly used a hashmap to store inorder indices for quick access, which is crucial for reducing the time complexity. The recursive helper function is appropriately defined with boundaries for the inorder array. One minor point: the variable name `pre_idx` is clear, but you might consider using a more descriptive name like `preorder_index` for consistency with `inorder_index`. However, this is a style choice and does not affect correctness. Overall, great job!

VERDICT: PASS

